Issue #59, opened 2021-10-13
Refactor social login to use common util to save data


lakhansamani:
**Version:** x.y.z

**Describe the bug**

**Steps To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

Issue #58, opened 2021-10-13
Allow login and token verification with multiple roles


lakhansamani:
**Version:** x.y.z

**Describe the bug**

**Steps To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

Issue #57, opened 2021-10-09
Social media signup should not support admin signups


lakhansamani:
**Version:** x.y.z

**Describe the bug**

**Steps To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

PR #60, opened 2021-10-13
feat: use multi roles login

lakhansamani:
#### What does this PR do?

- add support for protected roles
- refactor oauth code
- allow multi role login
- allow multi role verification


lakhansamani:
Resolves #57 
Resolves #58 
Resolves #59

PR #80, opened 2021-12-17
feat/add arangodb support

lakhansamani:
#### What does this PR do?

- Allow users to connect to the [ArangoDB](https://www.arangodb.com/) instance and have auth layer on top of that

- Add docs for `DATABASE_NAME` in env section. Default value is `authorizer` this is applicable to `arangodb` only


mysticaltech:
Your support for ArangoDB rocks! Thanks for adding that. So pleased to witness. Keep up the excellent work.

lakhansamani:
Thank you @mysticaltech

PR #88, opened 2021-12-28
Fix/app code spliting

lakhansamani:
#### What does this PR do?

- fix code splitting for app
- use default node + go setups for git actions
- reduces docker image size by remove node_modules
- optimize esbuild


lakhansamani:
Tested here: https://github.com/authorizerdev/authorizer/runs/4654343645?check_suite_focus=true

Issue #143, opened 2022-03-18
Add ability to activate account after revoked


lakhansamani:
**Feature Description**
There should be a way to activate revoked user account in dashboard.
This should be admin only feature.

**Describe the solution you'd like**

[ ] Update field in user schema `revoked_timestamp`. Set it to null
[ ] Add mutation called `_enable_access(param: {user_id: string})`
[ ] Add Option in dashboard to activate account
[ ] It should take confirmation before activating
[ ] It should allow login/signup for revoked access users  for all 3 recipes (basic, social, magiclink) 
[ ] Add docs for new API

**Describe alternatives you've considered**

**Additional context**

Issue #135, opened 2022-03-12
Add ability to revoke access from user


lakhansamani:
**Feature Description**
There should be a way to revoke access from user in dashboard. Technically known as soft delete of user.
This should be admin only feature.

**Describe the solution you'd like**

[ ] Add field in user schema `revoked_timestamp`
[ ] Add mutation called `_revoke_access(param: {user_id: string})`
      - It should set `revoked_timestamp` time in user schema
[ ] Add Option in dashboard to revoke
[ ] It should take confirmation before revoking
[ ] It should not allow login/signup for revoked access users for all 3 recipes (basic, social, magiclink) 
[ ] Add docs for new API

**Describe alternatives you've considered**

**Additional context**

PR #151, opened 2022-03-23
feat: update user access

anik-ghosh-au7:
#### What does this PR do?
Update user access from the dashboard.
Admin-only feature.

Describe the solution you'd like

[x] Add field in user schema revoked_timestamp
[x] Add mutation called _revoke_access(param: {user_id: string})
    - It should set revoked_timestamp time in user schema
[x] Add Option in dashboard to revoke
[x] Add mutation called _enable_access(param: {user_id: string})
[x] Add Option in the dashboard to activate account
[x] It should take confirmation before revoking
[x] It should take confirmation before activating
[x] It should not allow login/signup for revoked access users for all 3 recipes (basic, social, magiclink)
[ ] Add docs for new API

- https://github.com/authorizerdev/authorizer/issues/135
- https://github.com/authorizerdev/authorizer/issues/143

### DIFF
--- a/server/handlers/oauth_callback.go
+++ b/server/handlers/oauth_callback.go
@@ -95,7 +95,7 @@ func OAuthCallbackHandler() gin.HandlerFunc {
 			user.EmailVerifiedAt = &now
 			user, _ = db.Provider.AddUser(user)
 		} else {
-			if user.RevokedTimestamp != 0 {
+			if user.RevokedTimestamp != nil {
 				c.JSON(400, gin.H{"error": "user access has been revoked"})
 			}
 
--- a/server/handlers/oauth_callback.go
+++ b/server/handlers/oauth_callback.go
@@ -95,9 +95,12 @@ func OAuthCallbackHandler() gin.HandlerFunc {
 			user.EmailVerifiedAt = &now
 			user, _ = db.Provider.AddUser(user)
 		} else {
+			if user.RevokedTimestamp != 0 {
+				c.JSON(400, gin.H{"error": "user access has been revoked"})
+			}
+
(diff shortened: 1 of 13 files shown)


server/resolvers/enable_access.go
```diff
@@ -0,0 +1,44 @@
...
+	if !token.IsSuperAdmin(gc) {
+		return res, fmt.Errorf("unauthorized")
+	}
+
+	user, err := db.Provider.GetUserByID(params.UserID)
+	if err != nil {
+		return res, err
+	}
+
+	user.RevokedTimestamp = 0
```
lakhansamani, line 31: this should be set to `nil`

server/resolvers/enable_access.go
```diff
@@ -0,0 +1,44 @@
...
+
+	user, err = db.Provider.UpdateUser(user)
+	if err != nil {
+		log.Println("error updating user:", err)
+		return res, err
+	}
+
+	res = &model.Response{
```
lakhansamani, line 39: Message could be: `user access enabled successfully`

server/handlers/oauth_callback.go
```diff
@@ -95,9 +95,12 @@ func OAuthCallbackHandler() gin.HandlerFunc {
 			user.EmailVerifiedAt = &now
 			user, _ = db.Provider.AddUser(user)
 		} else {
+			if user.RevokedTimestamp != 0 {
```
lakhansamani, line 98: this should be `user.RevokedTimestamp != nil`

lakhansamani (changes requested):
Test cases are missing for this resolvers

Issue #157, opened 2022-04-21
Update Dashboard UI


lakhansamani:
**Feature Description**

Currently Dashboard UI is not user friendly. It has following issues

1. It is not responsive
2. Env variables are not categorized, as a result one has to scroll through lot of env vars
3. It does not have proper information on user page (Eg why is invite member disabled)
 

**Describe the solution you'd like**
1. Make login page responsive
2. Make main Dashboard layout responsive (one that appears after login)
3. Add tool tip to invite members button on users page on why it is disabled
4. Divide env page into following sub menu
    - OAuth Setting (Client ID, Client Secret)
    - Social Media Login
    - Roles
    - JWT Secrets
    - Session Storage
    - Email Configuration
    - Domain White Listing
    - Organization Information
    - UI Customization
    - Features
    - Database
    - Security (Admin Secret)

**Describe alternatives you've considered**

**Additional context**

lakhansamani:
Resolved https://github.com/authorizerdev/authorizer/releases/tag/0.23.0

PR #158, opened 2022-04-23
Update Dashboard UI 

akash-dutta-au7:
#### What does this PR do?
       The Login page is now responsive. 
       The main Dashboard layout is responsive now and the Env variables are categorized in different components.
       Tooltip added to the "Invite members" button on users page to inform why it is disabled.
         [Issue](https://github.com/authorizerdev/authorizer/issues/157)


lakhansamani:
This is replication of #161 and not needed any more

Issue #164, opened 2022-05-12
Unable to modify roles via dashboard or API


Vicg853:
**Version:** 0.22.0

**Describe the bug**
Errors with "invalid list of roles" although the inputted list is valid.

**Steps To Reproduce**
Simply by entering the dashboard and trying to append any user a role. Same behavior via API

- Update role mut
![image](https://user-images.githubusercontent.com/37887254/168135626-48ddadf2-9dcf-4fbf-a1c0-26d37776160a.png)
- Roles list query
![image](https://user-images.githubusercontent.com/37887254/168135888-5d8f8cc0-4cb4-49eb-be1e-72bec4a0258f.png)

**Desktop (please complete the following information):**

- OS: Windows (with WSL)
- Browser chrome  v101
- Version 0.22.0

Vicg853:
Sorry for introducing a new issue inside another one. But they are mostly related.

I don't know if this is on purpose, but: shouldn't we be able to remove all roles from a user?

Currently, you can't do it, as it ignores passing through the update roles [handling](https://github.com/authorizerdev/authorizer/blob/773213e5a45ec51c07342743c946c26790d3e65e/server/resolvers/update_user.go#L129) when the role list is empty. 

I tried doing the following to fix it: ([here](https://github.com/authorizerdev/authorizer/blob/773213e5a45ec51c07342743c946c26790d3e65e/server/resolvers/update_user.go#L147))
```go
if rolesToSave != "" || len(params.Roles) == 0 {
      user.Roles = rolesToSave
}
```
And it worked!
But I never used Go, so maybe there is a better solution or practice to follow.

lakhansamani:
@Vicg853 we don't want to assign empty roles to the user because it can mess up the authorization system, where roles are crucial. While signing up if no roles are passed `DEFAULT_ROLE` is assigned to the user.

lakhansamani:
@Vicg853 I was able to update roles from dashboard.
I also tested them after restarting server, they were persisted.

Can you share how are setting up ROLES env, is it via `.env` file or via `Dashboard` only?
JFYI: `.env` file variables are given higher preference

<img width="1440" alt="Screenshot 2022-05-13 at 7 57 17 AM" src="https://user-images.githubusercontent.com/6964334/168199312-26566bf9-bb5e-45ae-aa34-62e446845edf.png">

Also tested the JWT token generated after updating role

[...]

PR #165, opened 2022-05-12
Unable to update user role fix

Vicg853:
#### What does this PR do?
Fixes inverted: ``userRoles`` by ``roles``, under utils/validator -> ``IsValidRoles`` func. 

The function was iterating over all env roles and checking if each one existed under the newly inputted roles, causing it to reject, as it should do the opposite. 

Also, I added a comment on the [server/test/update_user_test.go](https://github.com/authorizerdev/authorizer/commit/4ceb6db4ba4920514f09866fe881ce8021ec1e2e#diff-1bf6c2931a0d52dee22f3081290af75a6762bfc5930d470fff7b968debd556c6), explaining the reason behind test error and that in reality succeeds.

#164

### DIFF
--- a/server/resolvers/update_user.go
+++ b/server/resolvers/update_user.go
@@ -26,6 +26,7 @@ func UpdateUserResolver(ctx context.Context, params model.UpdateUserInput) (*mod
 	if err != nil {
 		return res, err
 	}
+	fmt.Println(token.IsSuperAdmin(gc))
 
 	if !token.IsSuperAdmin(gc) {
 		return res, fmt.Errorf("unauthorized")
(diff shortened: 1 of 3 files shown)


lakhansamani:
@Vicg853 thank you for the fix

lakhansamani:
@Vicg853 I have updated test here: https://github.com/authorizerdev/authorizer/commit/ab717d956a47805d5a433982aaf48c4f676be12c

Thanks for catching that up.

lakhansamani:
@Vicg853 did a patch release here
https://github.com/authorizerdev/authorizer/releases/tag/0.22.1

PR #166, opened 2022-05-15
Fix/role update

Vicg853:
#### What does this PR do?
Fixes mistakes made on my previous pr

Also with vscode's help I found out about the two following lines that use this function [magic_link_login](https://github.com/authorizerdev/authorizer/blob/ab717d956a47805d5a433982aaf48c4f676be12c/server/resolvers/magic_link_login.go#L55) and [oauth_login](https://github.com/authorizerdev/authorizer/blob/ab717d956a47805d5a433982aaf48c4f676be12c/server/handlers/oauth_login.go#L55). 
So I fixed them too. 
Both manual and make tests run successfully 

#164 #165

### DIFF
--- a/server/handlers/oauth_login.go
+++ b/server/handlers/oauth_login.go
@@ -52,7 +52,7 @@ func OAuthLoginHandler() gin.HandlerFunc {
 
 			// use protected roles verification for admin login only.
 			// though if not associated with user, it will be rejected from oauth_callback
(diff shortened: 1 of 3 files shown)


lakhansamani (approved):

lakhansamani:
@Vicg853 thank you for the fix

PR #174, opened 2022-05-28
fix: replica cache consistency

### LINKED ISSUE #170: Updating env does not update cache on other replicas
**Version:** 0.22.2

**Describe the bug**

Current implementation

- On server start persist the current env state in db configured via env var.
- On server start put the env state in-memory for faster access
- 1st scenario helps in configuring env state via dashboard. Though I don't allow updating database related env via dashboard else it will be chicken-egg-problem :-D

There is a problem where updating env via dashboard will also not update the env var cache on other replicas and might require a restart for it.

**Steps To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

lakhansamani:
#### What does this PR do?

Fixes #170


lakhansamani:
@anik-ghosh-au7 thank you for the quick eye

PR #247, opened 2022-09-28
fix(server): use sameSite as lax by default for app cookie

lakhansamani:
#### What does this PR do?

Use sameSite = lax by default
Since app cookie can come from cross site it becomes important to set this in lax mode.
Example person using custom UI on their app domain and making request to authorizer domain.
For more information check:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
https://github.com/gin-gonic/gin/blob/master/context.go#L86


anik-ghosh-au7 (approved):
LGTM

Issue #272, opened 2022-10-11
Roles picker to multi select


jerebtw:
**Feature Description**

The current role picker is pretty annoying when you have multiple roles (5+ or so).
So it would be great to replace it with a multiple selection (for example https://mantine.dev/core/multi-select/).

And maybe even change the dashboard from chakra to mantine

lakhansamani:
Resolved and will be released as part of 1.1.22

PR #280, opened 2022-10-23
feat: add user roles multi select input

anik-ghosh-au7:
#### What does this PR do?
- add multi-select options to set user roles in `EditUserModal`

- resolves issue #272


anik-ghosh-au7:
@lakhansamani TBR

PR #286, opened 2022-11-03
fixed validation of refresh token

Pjort:
#### What does this PR do?
It breaks out of the else statement if refresh token fails to validate.

Fixes the problem were a invalid refresh token can be used to generate new good tokens

Big dislaimer though, I have no experience with GO and I just saw how the same logic was made for sessionKey, and noticed there was no return statements. I haven't tested the code, since I have no clue how to build in GO lang.


lakhansamani (approved):

lakhansamani:
@Pjort thanks for contributing, I will test more on my end.
Though can you share using which endpoints/apis you were able to use refresh token for multiple times

Issue #242, opened 2022-09-26
CouchDB Integration


VladimirCores:
CouchDB already has quite a lot of option to deal with auth, recently they added JWT (gateway) and only missing part is a issuer of this token. Please add. 
https://docs.couchdb.org/en/main/api/server/authn.html#:~:text=JWT%20authentication%20enables%20CouchDB%20to,JWT's%20%E2%80%9CNONE%E2%80%9D%20algorithm).

lakhansamani:
@VladimirCores thank you for reporting it.
I will soon try to integrate this.

manojown:
@lakhansamani I'm picking the couchbase provider integration.

lakhansamani:
Thanks @manojown

PR #287, opened 2022-11-04
Feat/couchbase db support

manojown:
#### What does this PR do?
DynamoDb Support - #242

  Providing the couchbase support

   NO
#### Env Require

1.  DATABASE_URL - Couchbase DB URL
2. DATABASE_USERNAME - DB User name
3. DATABASE_PASSWORD - DB Password
4. COUCHBASE_BUCKET - Couchbase Bucket name
5. COUCHBASE_SCOPE -  Couchbase Scope name or that can be `_default` as well
6. DATABASE_TYPE - `couchbase`


lakhansamani:
@manojown this PR will need a newly introduced db provider method called `GetUserByPhoneNumber`

https://github.com/authorizerdev/authorizer/blob/1eb8965f98dcea421ab99b5626b46afe930b9c79/server/db/providers/providers.go#L22

lakhansamani:
[discussion]:
What if we set by default and not allow this env vars
`COUCHBASE_BUCKET` -> `authorizer`
`COUCHBASE_SCOPE` -> `authorizer`

Any side effects you see for them?

manojown:
@lakhansamani  There is a problem if someone already has a database set up with some scope and wants to switch to an authoriser. In this case, they must make changes to their database and need to create a default scope. WYT?

manojown:
> @manojown this PR will need a newly introduced db provider method called `GetUserByPhoneNumber`
> 
> 
> 
> https://github.com/authorizerdev/authorizer/blob/1eb8965f98dcea421ab99b5626b46afe930b9c79/server/db/providers/providers.go#L22
> 
> 
> 
> 
Ok, I'll add this.

manojown:
@lakhansamani Is there any additional effect of `GetUserByPhoneNumber`, i.e. anything else I could have missed?

lakhansamani:
@manojown no usage of `GetUserByPhoneNumber` is already covered.

PR #288, opened 2022-11-07
Added login user to get all allowed roles as param

Pjort:
#### What does this PR do?
This makes it so that if a login request is made with roles params = "all_allowed_roles" then it will just return a JWT with all the users allowed roles

Currently you have to login once to know the allowed roles of a user, then send another request with the allowed roles added to the roles param in the request. This would make it possible to make one request and get a JWT that automatically holds all the users allowed roles in the roles claim.

DISCLAIMER: Again I am not a GO programmer, so I have no clue how to locally test 😕


Pjort:
After discussing it we came to the conclusion it was too specific to my use case, and it was possible to get the same effect by added the following to custom scripts: 
function(user,tokenPayload){var data = tokenPayload; data.role = user.roles; return data;}

PR #293, opened 2022-11-16
fix: codeVerifier shouldn't be empty for basic auth

luclu7:
#### What does this PR do?
When you need to get the clientID and clientSecret from the basic auth, codeVerifier SHOULD be set (so != "").

It fixes the OpenID flow for software that use POST forms (per example, Peertube, thanks @skid9000).


lakhansamani (approved):

lakhansamani:
@luclu7 @skid9000 thank you for sharing the fix,
I will release this as part 1.1.28-rc.3

PR #308, opened 2022-12-22
change toast position to top-right

LeanderGangso:
#### What does this PR do?
Moves toast(notification) to top-right.
Was previously bottom-right.

Found it annoying that it often blocked the save button which
is also at the bottom-right of the page.


lakhansamani (approved):
@LeanderGangso Thank you for creating this PR.
I agree with the user experience part here.

This change will be part of upcoming release.

Issue #345, opened 2023-04-17
Returning the IDS and Email of the invited users on invitation.


MussieT:
**Feature Description**

While inviting the members as an admin on `_invite_members`, I wanted to get the ids of the invited users on the response (it can be a list of ids or minimal user information like their email and id of the users who are invited).  I am unsure if this is a needed functionality from your side but was mandatory in my case. I wanted the user id and email for some business logic after they are invited.

I modified the mutation to return what I want on my local setup. Please let me know to submit a PR if this is a valid request.

lakhansamani:
@MussieT thank you for reporting will try to add in upcoming release

lakhansamani:
Released as part of #1.1.46

PR #346, opened 2023-04-19
Feat/return user info on invitation

MussieT:
#### What does this PR do?
Adds the functionality to return user information (id, email) on the invitation.

https://github.com/authorizerdev/authorizer/issues/345

https://github.com/authorizerdev/docs/compare/main...MussieT:docs:patch-1

**Preview**
Screencast 2023-04-19 16 09 28


lakhansamani (changes requested):
@MussieT can you also update the test for this particular resolver
https://github.com/authorizerdev/authorizer/blob/e822b6f31a24ca0a899115164119a26f78694edb/server/test/invite_member_test.go#L47

Overall it looks good to me

MussieT:
@lakhansamani Thanks for the response. But one question, the written test for the `invite_members` implementation checks whether the response is not `nil`. Should I modify it to check if the response contains both the `message` and `users`?

lakhansamani:
Yes that would be better test earlier response just had message now its good idea to check if it has expected users or not

MussieT:
@lakhansamani updated the test.

lakhansamani (approved):
LGTM

lakhansamani:
Thanks @MussieT

MussieT:
Appreciated the response @lakhansamani , planned to have more contributions soon.

PR #349, opened 2023-04-22
Features enhancement (Disable/Enable)

darshvaghela:

### DIFF
--- a/dashboard/src/components/EnvComponents/Features.tsx
+++ b/dashboard/src/components/EnvComponents/Features.tsx
@@ -8,86 +8,90 @@ const Features = ({ variables, setVariables }: any) => {
 		<div>
 			{' '}
 			<Text fontSize="md" paddingTop="2%" fontWeight="bold" mb={5}>
-				Disable Features
+				Features
 			</Text>
 			<Stack spacing={6}>
 				<Flex>
 					<Flex w="100%" justifyContent="start" alignItems="center">
-						<Text fontSize="sm">Disable Login Page:</Text>
+						<Text fontSize="sm">Login Page:</Text>
 					</Flex>
 					<Flex justifyContent="start">
 						<InputField
 							variables={variables}
 							setVariables={setVariables}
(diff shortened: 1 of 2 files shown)


dashboard/src/components/EnvComponents/Features.tsx
```diff
@@ -8,86 +8,90 @@ const Features = ({ variables, setVariables }: any) => {
...
 					<Flex w="100%" justifyContent="start" alignItems="center">
-						<Text fontSize="sm">Disable Login Page:</Text>
+						<Text fontSize="sm">Login Page:</Text>
 					</Flex>
 					<Flex justifyContent="start">
 						<InputField
 							variables={variables}
 							setVariables={setVariables}
 							inputType={SwitchInputType.DISABLE_LOGIN_PAGE}
+							is_Disable={true}
```
lakhansamani, line 23: 1. We should use camel-case (`isDisabled`) or snake-case (`is_disabled`). To maintain the consist I would recommend using camel-case.
2. is_Disabled name makes it feel as if the input is disabled as it is prop for input field. Instead, we should rename it to `hasReversedValue`.
3. You don't need to pass `true` value to boolean props by default if you just add `hasReversedValue` it will consider it true

dashboard/src/components/InputField.tsx
```diff
@@ -48,6 +48,7 @@ const InputField = ({
 	fieldVisibility,
 	setFieldVisibility,
 	availableRoles,
+	is_Disable,
```
lakhansamani, line 51: Pls update this to `hasReversedValue` and add the following comment

```js
// This prop is added to improve the user experience for the boolean ENV variable having `DISABLE_` prefix, as those values need to be considered in inverted form.
```

lakhansamani (changes requested):
Minor naming convention changes, rest looks good to me

lakhansamani (approved):
LGTM

PR #351, opened 2023-05-15
Feat/add sender name

miqe:
#### What does this PR do?
Adds the ability to customize the sender name displayed in the inbox, instead of just showing the email address.

This feature is important because it allows users to personalize the sender name that appears in the inbox of the recipient. By default, most email clients display the sender's email address as the sender name, which can be less recognizable to the recipient.


lakhansamani (approved):
Thanks @miqe this looks good to me

lakhansamani:
@miqe released as part of [1.1.48](https://github.com/authorizerdev/authorizer/releases/tag/1.1.48)

Issue #354, opened 2023-05-27
Facebook login is unable to insert (register) a user's email into the database. It is not properly registering a user's email into our database.


minilikmila:
**Version:** 1.1.46

**Describe the bug**
Third party authentication - Facebook login is unable to register user email to our DB. Although we configured the settings correctly and granted Facebook permissions for email and user profile, the login method is still unable to insert the user's email when they try to log in using Facebook.
In related to the previous issue, we have found that only the first user is able to successfully register, even if their email is not captured. This means that subsequent users attempting to register with Facebook are not being added to our user database,
Any subsequent users attempting to log in with Facebook are being logged in as the first user, with the access token being generated from that user's metadata.

Database - PostgreSQL

**Expected behavior**

To resolve the issue, we need to update the Facebook login handler so that it retrieves the user's metadata, including their email (based on the granted permission), and inserts this information into our user database. This will ensure that all users who log in with Facebook have their email captured and stored correctly, and that each user receives their own unique access token, rather than being logged in as the first user.

- OS:  Ubuntu
- Browser : chrome

lakhansamani:
Resolved with #355

PR #355, opened 2023-05-28
Modify the Facebook login authentication callback to enable user email access through the response body.

minilikmila:
#### What does this PR do?

The main change in this PR is the modification of the method to access the user's email returned from the Facebook response body during login.This fix resolves an issue with the previous method that caused an auth_callback error, and enables the Facebook login functionality to work as expected, allowing users to successfully log in and register their data.
Issue created in #354 is fixed under this PR.

#354

It doesn't affect any endpoint.


lakhansamani (approved):
LGTM,
Thanks for the contribution

Issue #360, opened 2023-06-14
Forgot password for users who signed up with their phone number


MussieT:
**Feature Description**

Users who sign up using their phone numbers should get an SMS message to reset their password.

**Describe the solution you'd like**

 1. The user sends forgot password mutation and receives sms code.
 2. The user enters the received code with a new password to change the password and resets it.

lakhansamani:
Implemented — ForgotPasswordRequest accepts phone_number, and the SMS-OTP reset path is fully wired end to end for phone-only signups. This has been supported since forgot_password's phone-number handling was added; no longer an open gap.

PR #364, opened 2023-06-26
Feat/forgot password via sms

MussieT:
#### What does this PR do?
This PR adds a `forget_password` implementation for users who signed up with their phone number to `reset_password` using their phone number according to the code they will be given.

#360 

https://docs.authorizer.dev/core/graphql-api

#### A bit of preview
forgot_

### DIFF
--- a/server/graph/schema.graphqls
+++ b/server/graph/schema.graphqls
@@ -418,15 +418,16 @@ input UpdateUserInput {
 }
 
 input ForgotPasswordInput {
-  email: String!
+  email_or_phone: String!
   state: String
   redirect_uri: String
 }
 
 input ResetPasswordInput {
-  token: String!
+  token_or_code: String!
   password: String!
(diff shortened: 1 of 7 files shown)


server/graph/schema.graphqls
```diff
@@ -418,15 +418,16 @@ input UpdateUserInput {
 }
 
 input ForgotPasswordInput {
-  email: String!
+  email_or_phone: String!
```
lakhansamani, line 421: Its better to keep `email` & add `phone_number`. Make both the fields optional and add validation in API that either of them should be present.

This will require less updates on SDK end and will also keep our APIs consistent

lakhansamani (changes requested):

lakhansamani:
This is resolved in https://github.com/authorizerdev/authorizer/releases/tag/1.3.6

PR #365, opened 2023-06-29
Fix wrong response_type parsed when missing response_mode

JokerQyou:
#### What does this PR do?

The original code incorrectly read configured default `respone_mode` value into a variable holding `response_type` value. This causes OAuth2 requests without `response_mode` passed via URL query parameter to fail, because `response_type` value will then be filled with illegal value.

There's no issue at this point. I was testing integration with [Memos](https://usememos.com) and found this coding typo. Their SSO implementation does not pass `response_mode`, thus resulting in error.

There's no API change.


JokerQyou:
@lakhansamani Please kindly review this PR and release a new version soon, thanks.

lakhansamani:
Thanks @JokerQyou for the contribution

lakhansamani:
@JokerQyou released as part of 1.1.52 (https://github.com/authorizerdev/authorizer/releases/tag/1.1.52)

PR #393, opened 2023-09-09
refactor(server/utils): remove redundant nil check

Juneezee:
#### What does this PR do?

From the Go specification:

> "3. If the map is nil, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/xzuA7WNrkF4


lakhansamani (approved):

lakhansamani:
Thanks @Juneezee

PR #420, opened 2023-11-20
Feat: add screen_hint param in /authorize api for explicit signup redirection

scaletech-milan:
#### What does this PR do?
Enable signup redirection via screen_hint param in /authorize API

### DIFF
--- a/server/constants/oauth2.go
+++ b/server/constants/oauth2.go
@@ -18,5 +18,5 @@ const (
 	ResponseTypeIDToken = "id_token"
 
 	// Constant indicating the "signup" screen hint for customizing authentication process and redirect to a signup page.
-	ScreenHint = "signup"
+	ScreenHintSignUp = "signup"
 )
--- a/server/constants/oauth2.go
+++ b/server/constants/oauth2.go
@@ -16,4 +16,7 @@ const (
 	ResponseTypeToken = "token"
 	// For the Implicit grant of id_token, use response_type=id_token to include an identifier token.
 	ResponseTypeIDToken = "id_token"
+
+	// Constant indicating the "signup" screen hint for customizing authentication process and redirect to a signup page.
+	ScreenHint = "signup"
 )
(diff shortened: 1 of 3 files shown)


server/handlers/authorize.go
```diff
@@ -120,27 +121,33 @@ func AuthorizeHandler() gin.HandlerFunc {
...
 		} else {
-			loginState += "&nonce=" + nonce
+			authState += "&nonce=" + nonce
 			if err := memorystore.Provider.SetState(state, nonce); err != nil {
 				log.Debug("Error setting temp code", err)
 			}
 		}
 
-		loginURL := "/app?" + loginState
+		authURL := "/app?" + authState
```
lakhansamani, line 137: Good idea is to not put `?` & `#` in authURL use it or declare another variable which just holds path `/app`or `/app/signup` and use that variable in line 140, 144, 146

lakhansamani (changes requested):
This PR is missing change in signup page which should also include social media buttons

scaletech-milan:
> This PR is missing change in signup page which should also include social media buttons

server/handlers/authorize.go
scaletech-milan: Updated the changes and added social media buttons on signup page

server/constants/oauth2.go
```diff
@@ -16,4 +16,7 @@ const (
 	ResponseTypeToken = "token"
 	// For the Implicit grant of id_token, use response_type=id_token to include an identifier token.
 	ResponseTypeIDToken = "id_token"
+
+	// Constant indicating the "signup" screen hint for customizing authentication process and redirect to a signup page.
+	ScreenHint = "signup"
```
lakhansamani, line 21: Lets rename this const to `ScreenHintSignUp`
If there is something else available for screen hint then also we can accomodate

lakhansamani (approved):
Just a minor suggestion overall looks good

lakhansamani (approved):

server/constants/oauth2.go
scaletech-milan: Yes True,
Updated!

PR #435, opened 2023-12-05
fix unable to update app_data for user

siimsams:
#### What does this PR do?

Fix bug where i'm unable to update just app-data. 

When updating with just app-data i get "please enter atleast one param to update". Even though AppData is present in the request.


lakhansamani (approved):
LGTM

PR #445, opened 2024-01-12
Feat/totp reset after recovery code validation

### LINKED ISSUE #442: Enhancement: TOTP Reset After Recovery Code Validation
**Feature Description**
Whenever TOTP is used as an MFA feature :
- The user will be redirected to the redirect_uri screen when they use the recovery code to log in, but they will first need to be redirected to the TOTP scanner image screen.

So after recovery codes are successfully validated, the same process repeats itself, with the user coming and logging in using TOTP as MFA. 

**Describe the solution you'd like**
Resetting the totp secret key and recovery codes inside the database after the recovery codes have been properly validated will cause the user to be redirected to the totp scanner image screen the next time, when the new recovery code and secret will be displayed.

**Describe alternatives you've considered**
**Additional context**

scaletech-milan:
#### What does this PR do?
It will reset the TOTP secret and update the recovery codes in the database.
After that user will be redirected to scanner image and the same old TOTP flow will continue until all recovery codes have been used. 

Resolve https://github.com/authorizerdev/authorizer/issues/442

NA

### DIFF
--- a/server/authenticators/providers/providers.go
+++ b/server/authenticators/providers/providers.go
@@ -2,8 +2,6 @@ package providers
 
 import (
 	"context"
-
-	"github.com/authorizerdev/authorizer/server/graph/model"
 )
 
 // AuthenticatorConfig defines authenticator config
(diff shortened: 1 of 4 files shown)


lakhansamani:
Closing this PR — thanks for the contribution back in 2024.

The codebase went through a full restructure since this was opened (`server/...` → `internal/...`, config moved to CLI flags), so this PR now targets files that no longer exist (`server/authenticators/providers/totp/totp.go`, `server/resolvers/verify_otp.go`, etc.) and GitHub reports it as unmergeable due to conflicts. It can't be rebased forward as-is.

That said, the underlying request in #442 is still valid — recovery-code login today still doesn't reset the TOTP secret/regenerate recovery codes, so a user who logs in via recovery code keeps re-using the same (partially consumed) set indefinitely. I'm leaving that issue open for a fresh implementation against the current `internal/authenticators/totp` structure.

PR #467, opened 2025-04-21
check X-Forwarded-Host header

Aleksei-Badyaev:
#### Authorization in the intermediate layer of the backend, when the "authorizer" server is accessible from the frontend and backend at different URLs.


lakhansamani:
@Aleksei-Badyaev thank you for the PR we have resolved this v2 launch.

Just to avoid conflict closing this for now

PR #471, opened 2025-10-01
Feature, separate HttpOnly cookie attribute APP_COOKIE_HTTPONLY

localnerve:
#### What does this PR do?

* Adds APP_COOKIE_HTTPONLY env variable to govern the setting of the app cookie HttpOnly attribute separately from the Secure attribute (still set by APP_COOKIE_SECURE). APP_COOKIE_SECURE now only effects the Secure cookie attribute alone and not also the HttpOnly attribute.
* Fixes asymmetric SameSite discrepency when APP_COOKIE_SECURE=false and SameSite is LAX. (prior code kept Delete as SameSite=None, causing mismatch and non-deletes on client side).
* Fixes the testSuite to execute without crashing after prior API updates that result in correct err and nil responses (see test suite for updates, pretty minor, but necessary to keep test suite functional).

This does not directly address an issue. This allows webdevs to run authorizer instances in a test environment without setting up Https. This is done by APP_COOKIE_SECURE=false, and with this PR, now without disabling HttpOnly cookie attribute.
Why? **WEBKIT CLIENT TESTING**
This allows test and application code to run normally (expecting HttpOnly cookies) while allowing **webkit** clients to receive data without setting up **Https proxies** and further complicating the test environment. Webkit only allows cookies to be stored over non-TLS connections with Secure attribute disabled. This update allows webdevs to test with webkit clients without setting up TLS proxies on localhost like other browsers (firefox, chromium).

This PR adds another undocumented Environment Variable (alongside APP_COOKIE_SECURE) APP_COOKIE_HTTPONLY. This allows APP_COOKIE_SECURE to be disabled without effecting the HttpOnly attribute.
This PR adds makefile test `make test-debug` and documentation in CONTRIBUTING.md was updated to reflect this. I had to use this to debug/fix the existing tests.


lakhansamani:
Revisiting this as part of a post-2.4.0-rc.3 audit of open issues/PRs.

One of the two things this PR fixes has since been resolved independently: session-cookie deletion now takes the same \`sameSite\` parameter used when the cookie was set (\`internal/cookie/cookie.go\`'s \`DeleteSession\`/\`BuildDeleteSessionCookies\`), so the asymmetric-SameSite-on-delete bug this PR described no longer reproduces on current \`main\`.

The other part — a dedicated \`APP_COOKIE_HTTPONLY\` flag to decouple \`HttpOnly\` from \`APP_COOKIE_SECURE\` — is still not implemented; \`HttpOnly: true\` is still hardcoded in \`BuildSessionCookies\`/\`BuildDeleteSessionCookies\`. So this PR isn't fully superseded — not closing it, but it would need a rebase given how much \`internal/cookie/\` has changed since it was opened (SameSite is now a first-class parameter threaded through, MFA session cookies were added alongside the app session cookie, etc.).

Issue #508, opened 2026-03-30
RFC: Fine-Grained Permissions Model


lakhansamani:
## RFC: Fine-Grained Permissions Model

**Phase**: 2 — Authorization & M2M  
**Priority**: P1 — Critical  
**Estimated Effort**: High  
**Depends on**: Audit Logs (#505)

---

### Problem Statement

Authorizer only supports basic RBAC with comma-separated role strings on the User schema (`Roles` field). There is no permissions model, no resource-level access control, and no way to check "can user X perform action Y on resource Z?" WorkOS FGA, Keycloak Authorization Services, and Clerk all go beyond simple roles. This is the most requested enterprise feature for any auth platform.

---

### Current Architecture Context

- User schema has `Roles string` — comma-separated (e.g., `"user,admin"`)
- Roles configured via `--roles` and `--default-roles` CLI flags
- `--protected-roles` prevents self-assignment
- JWT claims include `roles` (string array) and `allowed_roles`
- `--jwt-role-claim` configures the claim name (default: `"role"`)
- No Permission, Resource, or RolePermission schemas exist
- GraphQL admin mutations `_update_user` can change roles
- Custom access token script (`--custom-access-token-script`) via Otto JS VM can add custom claims

---

#### 1. Data Model

**New schemas** in `internal/storage/schemas/`:

```go
// Permission represents a named permission (e.g., "documents:read")
type Permission struct {
    ID          string `json:"id" gorm:"primaryKey;type:char(36)"`
    Name        string `json:"name" gorm:"type:varchar(100);uniqueIndex"`   // e.g., "documents:read"
    Description string `json:"description" gorm:"type:text"`
    CreatedAt   int64  `json:"created_at" gorm:"autoCreateTime"`
    UpdatedAt   int64  `json:"updated_at" gorm:"autoUpdateTime"`
}

// RolePermission maps roles to permissions (many-to-many)
type RolePermission struct {
    ID           string `json:"id" gorm:"primaryKey;type:char(36)"`
    Role         string `json:"role" gorm:"type:varchar(100);index:idx_rp_role;uniqueIndex:idx_rp_unique"`
    PermissionID string `json:"permission_id" gorm:"type:char(36);index:idx_rp_permission;uniqueIndex:idx_rp_unique"`
    CreatedAt    int64  `json:"created_at" gorm:"autoCreateTime"`
}

// Resource represents a protected resource instance
type Resource struct {
[...]

PR #607, opened 2026-04-14
feat: Fine-Grained Authorization with Keycloak-Core Policy Engine

lakhansamani:
## Summary

Implements RFC #508 — Fine-Grained Authorization with a Keycloak-inspired four-pillar model (Resources, Scopes, Policies, Permissions). Replaces flat comma-separated role strings with a composable, principal-agnostic authorization engine that's **always enforcing** and opt-in per call.

## How callers consume it

Three GraphQL operations gained an optional `required_permissions: [PermissionInput!]` field — AND semantics, any deny or unmatched `(resource, scope)` returns `unauthorized`:

- `session`
- `validate_session`
- `validate_jwt_token`

Pre-existing callers that omit the field see no behavior change.

Authenticated users can list their own grants via the `my_permissions` query. Admins manage the policy graph via `_add_resource` / `_add_scope` / `_add_policy` / `_add_permission` (plus list / update / delete for each).

## Key features

- **Resources / Scopes / Policies / Permissions** — composable four-pillar model
- **Decision strategies** — `affirmative` (any policy grants) and `unanimous` (all must grant); explicit deny always wins
- **Principal-agnostic evaluator** — supports users, M2M clients, AI agents via a typed `Principal` abstraction with optional `MaxScopes` delegation ceiling
- **In-memory caching with negative caching** — invalidated automatically on admin CRUD; TTL via `--authorization-cache-ttl` (default 300s, 0 to disable)
- **Dashboard UI** — 4 pages (Resources, Scopes, Policies, Permissions) with inline help text + natural-language permission summaries
- **Policy target validation** — role-target values must match a configured `--roles` value at create/update time so a typo cannot silently produce a dead policy
- **Always enforcing** — no permissive "log but allow" mode; the previously-proposed `--authorization-enforcement` flag and its dual modes were dropped before shipping

## CLI flags added

- `--authorization-cache-ttl` (int, default `300`) — cache TTL in seconds; `0` disables.
- `--include-permissions-in-token` (bool, default `false`) — embed grants in JWT access tokens.
- `--authorization-log-all-checks` (bool, default `false`) — audit-log every check, not just denials.

No CLI flag controls enforcement — authorization is unconditionally enforcing.

## Observability

Two Prometheus counter families plus a histogram:

| Metric | Labels | Purpose |
|---|---|---|
| `authorizer_authz_checks_total` | `result` (`allowed\|denied\|unmatched\|error`) | Every `CheckPermission` call. |
| `authorizer_authz_unmatched_total` | — | Subset that found no permission row for `(resource, scope)`. Use to find policy-graph gaps. |
| `authorizer_authz_check_duration_seconds` | — | End-to-end evaluator latency. |
| `authorizer_required_permissions_checks_total` | `endpoint` (`session\|validate_session\|validate_jwt_token`), `outcome` (`granted\|denied\|not_requested\|error`) | Per-endpoint FGA adoption + enforcement signal. **Alert on `outcome="error"`** — should sit at zero. |

Startup probe emits `authz: 0 permissions configured — all authorization checks will DENY` when the database has zero permissions, so operators don't lock themselves out silently.

Authz CRUD operations (add/update/delete for permission/policy/resource/scope) emit audit-log entries for compliance.

## Storage

- 4 new tables (`authz_resources`, `authz_scopes`, `authz_policies`, `authz_permissions`) and 3 join tables, replicated across **all 13 storage providers** (PostgreSQL, MySQL, SQLite, SQL Server, YugabyteDB, MariaDB, PlanetScaleDB, CockroachDB, LibSQL, MongoDB, ArangoDB, Cassandra/ScyllaDB, Couchbase, DynamoDB).
- 28 new methods on the `storage.Provider` interface — full CRUD plus the optimized `GetPermissionsForResourceScope` join used on the evaluator hot path.

## Backward compatibility

- `User.Roles`, `IsSuperAdmin()`, JWT format — **unchanged**.
- `required_permissions` is per-call opt-in. Callers that omit it preserve pre-FGA semantics exactly.
- Permissions are added to JWT only when `--include-permissions-in-token=true`.

## Future-proof for M2M and AI agents

`Principal{ID, Type, Roles, MaxScopes}` accepts `type: "user" | "client" | "agent"`. Adding a new principal type is one switch case, no schema migration. `MaxScopes` is a delegation ceiling: even if policies grant more, scopes outside the ceiling are denied.

## Security guards

- Negative caching of unmatched checks with a bounded counter to prevent unbounded cache growth from attacker-controlled `(resource, scope)` input reaching authenticated GraphQL.
- Fail-closed on storage probe error in `validateResourceExists` — a transient DB blip surfaces as an error to the caller rather than as `Allowed: true`.
- Role-aware cache key so cross-tenant `(user, resource, scope)` results cannot collide.
- Atomic `UpdatePermission` with compensating rollback on failure mid-write.

## Test plan

- [x] **`make test` green** — every package passes against SQLite, including `internal/authorization`, `internal/graphql`, `internal/metrics`, `internal/storage`, and the full `internal/integration_tests` suite (~54s).
- [x] **`internal/storage` package** — `TestStorageProvider` covers every storage method on SQLite (Resource / Scope / Policy / Permission CRUD, the `GetPermissionsForResourceScope` join, MFA Session, Audit Log, OAuth State, etc.).
- [x] **FGA integration tests** — `TestAuthorizationCRUD` (full CRUD on resources/scopes/policies/permissions + role-based grant check + admin-role deny check), `TestCheckPermission_NoPermissions_Denies`, `TestCheckPermission_ExplicitDenyPolicy_Denies`, `TestCheckPermission_ExplicitDenyOverridesAffirmativeGrant`, `TestCheckPermission_CacheKeyIncludesRoles`, `TestUpdatePermission_InvalidScopeDoesNotDropExistingLinks`, `TestAddPermission_DuplicateNameReturnsConflict`, `TestCheckPermission_IncrementsPrometheusCounters`, `TestCheckPermission_UnknownResource_DeniesAndDoesNotBumpUnmatchedCounter`, `TestCheckPermission_ResultLabels_IncrementCorrectCounter`, `TestCheckPermission_MaxScopes_*`, `TestCheckPermission_UnanimousDecisionStrategy_AllPoliciesMustAgree`, `TestCheckPermission_UserTypePolicy_MatchesOnPrincipalID`.
- [x] **`required_permissions` integration test** — `TestRequiredPermissions` covers backward-compat + granted + denied paths for `session`, `validate_session`, `validate_jwt_token`, plus `metrics counters increment per outcome` assertion that verifies the new counter increments on `granted`, `denied`, and `not_requested`.
- [x] **Policy-target validator** — `TestValidatePolicyTargets` (6 cases: valid role / valid user / mismatched target_type / unknown role / empty value / empty targets).
- [x] `go build ./...` clean.
- [x] `go vet ./...` clean except a pre-existing mongodb context-leak hint unrelated to this branch.
- [x] Dashboard `npm run build` passes.
- [ ] Multi-DB storage tests via `make test-all-db` (Docker, runs the same `TestStorageProvider` suite across postgres / sqlite / mongodb / arangodb / scylladb / dynamodb / couchbase). Pending.
- [ ] Manual smoke: `make dev`, log into dashboard, define a resource → scope → policy → permission, then call `validate_jwt_token` with `required_permissions: [{resource, scope}]` and confirm granted / denied / unauthorized responses match policy intent.

## Docs follow-up

Companion PR against `authorizer-docs` adds `core/authorization.md` (model reference, admin mutations, `required_permissions` usage, decision strategies, observability), updates `core/graphql-api.md` for the new field + `my_permissions` query, refreshes `core/metrics-monitoring.md` with the new counter, and adds an FGA section to `migration/v1-to-v2.md`. Plan written; PR opens after this lands.

## Design specs

- `docs/superpowers/specs/2026-04-13-fine-grained-authorization-design.md` — original FGA design


lakhansamani:
## Update: Authorization enforcement flag removed; new per-endpoint metric

Latest commits on this branch (`b3ef1c33..2aa818d0`) finish the FGA enforcement story by **removing the dual-mode behavior entirely** in favor of always-enforcing semantics, replacing the rollout-safety promise of the legacy `--authorization-enforcement` flag with structured per-endpoint observability.

### Why
The flag was a footgun: its default (`permissive`) silently allowed `required_permissions` checks against unmatched `(resource, scope)` pairs, with only a rate-limited warn log as feedback. Since `required_permissions` is opt-in per call, the rollout pain it was solving never existed — callers either ask for the check (and want it enforced) or skip the field entirely (and see no change). Fail-open as the default mode of a security primitive was the wrong default.

### What changed
- **`--authorization-enforcement` deprecated**: still parses (so existing systemd/Docker configs don't break) but its value is ignored. Operators see Cobra's built-in deprecation notice plus a structured `Warn` log on startup if they pass it.
- **Evaluator collapsed to enforcing-only**: `handleNoPermission` now has a single deny path; `warn_limiter.go` deleted.
- **Prometheus label cleanup**: `mode` label removed from `authorizer_authz_checks_total` and `authorizer_authz_unmatched_total`; the result values `unmatched_allowed`/`unmatched_denied` collapse into `unmatched`. **Dashboards and recording rules need updating** — see `MIGRATION.md`.
- **New counter `authorizer_required_permissions_checks_total{endpoint, outcome}`**: per-endpoint FGA adoption + denial signal. Outcomes are `granted`, `denied`, `not_requested`, `error`. Endpoints are `session`, `validate_session`, `validate_jwt_token`. Alert on `outcome="error"`.
- **Tests**: `testSetupWithAuthzMode` helper deleted; permissive-mode test scenarios pruned; `NormalizeAuthzEnforcement` and its 5 unit tests removed; one test (`TestCheckPermission_PermissiveDefault_NoPermissions_Allows`) deleted since its premise is gone, another renamed to drop the no-longer-meaningful "Permissive" qualifier.
- **Constants `AuthorizationEnforcementPermissive`/`Enforcing`** deleted.
- **Docs**: `CHANGELOG.md` "Breaking changes" entry; `MIGRATION.md` gains an "Authorization Enforcement Removal" section with the pre-upgrade audit checklist (Prometheus `authorizer_authz_unmatched_total` + `authz.unmatched=true` logs), the dashboard relabel guidance, and the new metric's PromQL examples.

### Test plan
- [x] Full `make test-sqlite` green (every package PASS including all `TestCheckPermission_*`, `TestRequiredPermissions`, `TestMetrics`).
- [x] `go build ./...` clean; `go vet ./...` clean except a pre-existing mongodb context-leak hint that predates this branch.
- [x] Manual deprecation smoke-test: server started with `--authorization-enforcement=permissive` emits the cobra deprecation notice on first line of stderr.
- [x] Manual `required_permissions` smoke: `required_permissions: [{resource: "nonexistent", scope: "read"}]` against an undefined pair now returns `unauthorized` (was silently `is_valid: true` under the old default).
- [x] Metrics smoke: `curl /metrics | grep required_permissions_checks_total` shows the new counter with `endpoint` + `outcome` labels.

### Companion docs PR (separate repo)
The `authorizer-docs` repo needs the matching reference page (`core/authorization.md`), GraphQL API updates (`required_permissions` on session APIs, `my_permissions` query, admin FGA mutations), metrics page additions, and the v1→v2 FGA upgrade checklist. Plan saved locally; the docs PR will land separately so docs accurately describe shipped behavior.

PR #618, opened 2026-05-30
test+fix: address principal-engineer review findings (Phases 0-4)

lakhansamani:
> **Stacked on top of #617 (which is stacked on #616 → #615 → #614).** Review and merge after Phase 4.

## Summary
Adds the **missing integration/e2e test coverage** flagged across reviews of #614 / #615 / #616 / #617, and **fixes the blocking bugs** those reviews surfaced. Test additions are spread across every layer the PR stack introduces; fixes are deliberately scoped to issues that would have hit production on the first deploy.

## New tests (8 files, ~50 sub-cases)

| File | Coverage |
|---|---|
| \`internal/parsers/url_test.go\` | \`GetHostFromRequest\` priority + spoof rejection; \`GetAppURLFromRequest\` |
| \`internal/cookie/cookie_test.go\` | \`BuildSessionCookies\`/\`BuildMfaSessionCookies\` shape, MaxAge, Secure, HttpOnly, SameSite; \`ParseSameSite\` |
| \`internal/service/sideeffects_test.go\` | \`MetaFromGin\`/\`ApplyToGin\` nil-safety + cookie round-trip |
| \`internal/grpcsrv/interceptors/interceptors_test.go\` | Recovery (panic→Internal), Logging (per-code level), Validate (protovalidate) |
| \`internal/grpcsrv/transport/grpc_metadata_test.go\` | \`MetaFromGRPC\` host/IP/UA/auth/cookies + \`:authority\` fallback |
| \`internal/mcp/schema_test.go\` | Flat scalars, nested message, **cycle-safety regression test for AppData→Struct→Value** |
| \`internal/integration_tests/grpc_surface_test.go\` | All 18 stub RPCs verified to return \`codes.Unimplemented\` + gRPC health endpoint |
| \`internal/integration_tests/rest_openapi_test.go\` | \`/openapi.json\` serves valid swagger 2.0 |
| \`internal/integration_tests/mcp_stubs_test.go\` | MCP stubbed-tool call returns \`CallToolResult{IsError:true}\` |

## Blocking fixes from review

| # | Where | Fix |
|---|---|---|
| #616-1 | \`cmd/root.go\` | **--grpc-port** default 8081 collided with **--metrics-port**. Moved gRPC to **9091**. Added triple-port collision check (HTTP/metrics/gRPC); fails fast at startup. |
[...]

PR #619, opened 2026-05-30
refactor(proto): consolidate to single Authorizer service; method names mirror GraphQL ops

lakhansamani:
> **Stacked on top of #618.** Per your direction: collapse the nine per-resource proto services into one **Authorizer** service whose method names match the GraphQL operation names 1:1.

## What changed
The 9 services (Meta / User / Session / MagicLink / EmailVerification / PasswordReset / OtpChallenge / Token / Authz) become **one Authorizer service with 19 RPCs**:

\`Signup\`, \`Login\`, \`Logout\`, \`MagicLinkLogin\`, \`VerifyEmail\`, \`ResendVerifyEmail\`, \`VerifyOtp\`, \`ResendOtp\`, \`ForgotPassword\`, \`ResetPassword\`, \`Profile\`, \`UpdateProfile\`, \`DeactivateAccount\`, \`Revoke\`, \`Session\`, \`ValidateJwtToken\`, \`ValidateSession\`, \`Meta\`, \`Permissions\`

## Trade-offs noted
**Pros:** mirrors GraphQL surface users already know; single typed client per language; no resource-name conventions to learn.
**Cons (vs the AIP-121/122 design we had):** loses \`List/Get/Create\` symmetry per resource; doesn't scale as cleanly past 50+ methods. Acceptable for an auth surface where most ops are stateless verbs and admin services stay on GraphQL.

## Design decisions called out
1. **Buf STANDARD's \`RPC_REQUEST_RESPONSE_UNIQUE\` stays enforced.** Methods whose underlying payload is shared (\`AuthResponse\`, \`User\`, \`Meta\`) wrap it in per-RPC \`SignupResponse\`/\`LoginResponse\`/\`MetaResponse\`/etc. One extra accessor on the client side; one extra field per response — but lint passes and each response can evolve independently.
2. **\`SERVICE_SUFFIX\` lint excepted** in \`buf.yaml\`. The single service is intentionally named \`Authorizer\`, not \`AuthorizerService\` — it IS the authorizer.
3. **REST mapping rule:** \`GET /v1/{method}\` for genuinely-empty queries (Meta, Profile, Permissions, Logout); \`POST /v1/{method}\` for everything else. Path uses snake_case method name (\`POST /v1/magic_link_login\`, \`/v1/forgot_password\`, ...) so REST clients see the same identifier they'd see in GraphQL.
4. **MCP-exposed tools** (proto annotation \`mcp_tool.exposed=true\`): \`meta\`, \`profile\`, \`session\`, \`permissions\`. Credential-bearing methods stay unexposed.

## Files
- **Deleted**: \`proto/authorizer/{meta,user,session,verification,token,authz}/v1/**\` and their generated counterparts; \`internal/grpcsrv/handlers/{meta,stubs}.go\`.
- **New**: \`proto/authorizer/v1/{authorizer,types}.proto\`, \`gen/go/authorizer/v1/**\` regenerated, \`internal/grpcsrv/handlers/authorizer.go\` (Meta real, 18 stubs).
- **Updated**: \`proto/buf.yaml\`, \`internal/grpcsrv/server.go\`, \`internal/gateway/mount.go\`, and the 7 test files that referenced old per-service types.
- **Unchanged**: \`internal/mcp/{scanner,schema,server}.go\` (annotation-driven; discovered tools from any proto package without modification), \`internal/grpcsrv/transport/grpc_metadata.go\`, \`internal/service/\`, \`cmd/root.go\`.

## Test plan
- [x] \`buf lint proto\` clean
- [x] All 17 packages green: integration suite 65s, no regression
- [x] \`TestAuthorizerStubsReturnUnimplemented\` locks down the 18 not-yet-migrated methods
- [x] \`TestMCPListAndCallMeta\` confirms 4 MCP tools (meta, profile, session, permissions) auto-discovered from annotations


lakhansamani:

PR #628, opened 2026-06-15
fix(storage): drop stale unique email/phone constraints on upgrade

lakhansamani:
## Problem

Upgrading an existing SQL database to v2 fails at startup with:

```
ERROR: constraint "uni_authorizer_users_email" of relation "authorizer_users" does not exist (SQLSTATE 42704)
ALTER TABLE "authorizer_users" DROP CONSTRAINT "uni_authorizer_users_email"
failed to create storage provider
```

v1 declared `email` and `phone_number` as **UNIQUE** on `authorizer_users` and `authorizer_otps`. v2 keeps plain (non-unique) indexes and enforces uniqueness in the application layer (`AddUser`/`UpdateUser`), so behaviour is identical across all 13+ databases and these fields can stay optional (email-only / phone-only signups).

On an upgraded DB the columns are still unique, so GORM `AutoMigrate` tries to `DROP CONSTRAINT "uni_<table>_<col>"` — a name that does **not** match what the database actually created — failing with 42704 and aborting the whole migration. It surfaces table-by-table (`authorizer_users`, then `authorizer_otps`, then the `phone_number` columns).

## Fix

Clear the legacy uniqueness for all four columns (`authorizer_users` / `authorizer_otps` × `email` / `phone_number`) **before** `AutoMigrate`. Depending on the v1 release the old uniqueness exists either as:

1. a **constraint** — `<table>_<col>_key` (Postgres/MySQL) or `uni_<table>_<col>` (GORM) → dropped by name via `DropConstraint`.
2. a standalone **unique index** — e.g. `idx_authorizer_otps_phone_number` → dropped name-agnostically via `GetIndexes`, only when the index is unique and single-column on email/phone, so the current non-unique indexes are left intact and `AutoMigrate` recreates anything it needs.

All operations are guarded (`HasConstraint`) and non-fatal — fresh installs have nothing to drop. Generalises the prior `phone_number`-only handling. Duplicate emails/phone numbers are still rejected (now by the server, not the DB).

## Notes

- Affects all SQL backends (Postgres/Yugabyte/CockroachDB/MySQL/MariaDB/SQLServer). SQLite and NoSQL backends are unaffected.
- Docs: companion migration-guide update in `authorizer-docs` (`docs/upgrade-unique-constraint-migration`).

## Test plan

- [x] `go build` / `go vet` pass
- [ ] Manual: seed a legacy unique constraint **and** a legacy unique index on `authorizer_users.email` / `authorizer_otps.phone_number`, then start the server and confirm migration succeeds and the columns become non-unique.

### DIFF
--- a/internal/storage/db/sql/provider.go
+++ b/internal/storage/db/sql/provider.go
@@ -74,12 +74,69 @@ func NewProvider(
 		return nil, err
 	}
 
-	// For sqlserver, handle uniqueness of phone_number manually via extra db call
-	// during create and update mutation.
-	if sqlDB.Migrator().HasConstraint(&schemas.User{}, "authorizer_users_phone_number_key") {
-		err = sqlDB.Migrator().DropConstraint(&schemas.User{}, "authorizer_users_phone_number_key")
-		if err != nil {
-			deps.Log.Debug().Err(err).Msg("failed to drop unique constraint on phone_number")
+	// v1 declared the email and phone_number columns of authorizer_users and
+	// authorizer_otps as UNIQUE. v2 keeps plain (non-unique) indexes and enforces
+	// uniqueness in the application layer (see AddUser/UpdateUser) so behaviour is
+	// identical across all supported databases and these fields can stay optional
+	// (email-only or phone-only signups).
+	//
+	// On an upgraded SQL database those columns are still unique, so GORM
+	// AutoMigrate tries to DROP CONSTRAINT "uni_<table>_<col>" — a name that does
+	// not match what the database actually created — failing with SQLSTATE 42704
+	// ("constraint does not exist") and aborting the whole migration. Depending on
+	// the v1 release, the old uniqueness exists either as a CONSTRAINT
+	// (<table>_<col>_key) or as a standalone UNIQUE INDEX (idx_<table>_<col>), so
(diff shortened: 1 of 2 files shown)


lakhansamani:
Verified against a real Postgres (the 2.3.0 GORM 1.25.10 stack), confirming the fix is required and not redundant:

**Unfixed (`main`)** — seeded a legacy `UNIQUE` constraint on `authorizer_users.email` and a legacy unique index on `authorizer_otps.phone_number`, then ran migration:

- raw `AutoMigrate` → `ERROR: constraint "uni_authorizer_users_email" ... does not exist (SQLSTATE 42704)`
- `NewProvider` (real entrypoint) → same `42704`

So GORM 1.25.10 does **not** reconcile/drop the stale unique objects on its own — startup deterministically aborts.

**With this branch** — both tests pass on Postgres:

```
TEST_DBS=postgres go test ./internal/storage/db/sql/ -run 'TestProviderEmailPhoneUpdatesAndUniqueness|TestStaleUniqueConstraintMigration'
--- PASS: TestProviderEmailPhoneUpdatesAndUniqueness/postgres
--- PASS: TestStaleUniqueConstraintMigration/postgres
```

The migration drops the stale uniqueness, the non-unique search index is preserved, and a post-migration phone-number update succeeds.

PR #635, opened 2026-06-16
feat(api)!: serve all auth ops on gRPC/REST + flatten response envelope

lakhansamani:
## Summary

Brings the gRPC/REST surface to full parity with GraphQL, on a single consolidated proto package, with the response envelope flattened to match GraphQL byte-for-byte.

### 1. All auth operations now work on gRPC + REST
The 10 remaining auth ops were GraphQL-only. Their business logic is moved into the transport-agnostic `internal/service` layer (single source of truth); GraphQL resolvers, gRPC handlers, and the REST gateway all delegate to it:

`login`, `magic_link_login`, `verify_email`, `resend_verify_email`, `verify_otp`, `resend_otp`, `forgot_password`, `reset_password`, `update_profile`, `deactivate_account`.

GraphQL resolvers are now thin wrappers — no duplicated logic.

### 2. Flattened response envelope (GraphQL parity)
Public RPCs return the bare domain message instead of a per-RPC wrapper:
- `signup`/`login`/`verify_email`/`verify_otp`/`session` → `AuthResponse` (was `{auth:{…}}`)
- `profile` → `User` (was `{user:{…}}`)
- `meta` → `Meta` (was `{meta:{…}}`)

Field names are snake_case (`UseProtoNames`) and verified identical to the GraphQL schema, so GraphQL/REST/gRPC payloads are now byte-identical.

### 3. Proto consolidation
`authorizer/common/v1/*` merged into a single `authorizer.v1` package. Regenerated Go, OpenAPI, and the Go/Python/TS client stubs.

### 4. Fix: REST gateway dropped cookies
Session/MFA cookies are emitted as `set-cookie` gRPC metadata. The gateway had no outgoing-header matcher, so grpc-gateway renamed them to `Grpc-Metadata-Set-Cookie` — browsers ignored them. Added a matcher that promotes them to real `Set-Cookie` headers (and stops other gRPC metadata leaking). Admin login/logout/session cookies over REST are fixed by the same change. GraphQL cookies (the `ApplyToGin` path) are untouched.

## Breaking changes
`/v1` REST + gRPC responses for the 7 flattened RPCs change shape (no `auth`/`user`/`meta` wrapper). `/v1` is unreleased, and the regenerated clients expose the flat types. Buf lint rules `RPC_RESPONSE_STANDARD_NAME` and `RPC_REQUEST_RESPONSE_UNIQUE` are waived in `buf.yaml` (documented) since responses intentionally reuse shared domain messages.

## Tests
- 10 migrated ops keep passing all existing GraphQL integration tests (behavior parity).
- New: end-to-end gRPC login + auth-required UpdateProfile; positive surface contract for all 10 RPCs.
- New: REST `Set-Cookie` regression tests for user signup and admin login.
- Full `internal/integration_tests` + `grpcsrv` + `gateway` + `service` suites green on SQLite.

Docs updated in `authorizer-docs` (separate PR): rest-api.md / grpc.md — flat response shape, cookie behavior, single proto package.


lakhansamani:
## Review: principal-engineer + security-engineer (adversarially verified)

Both passes diffed every migrated service file line-by-line against `origin/main` and ran build/vet/tests. The migration was verified faithful; two real defects were found and **fixed in 77ce4ca**.

### Must-fix (resolved)
1. **[High] `UpdateProfile` could silently disable MFA over gRPC/REST.** `model.UpdateProfileRequest.IsMultiFactorAuthEnabled` is `*bool` and the service keys disable on nil-vs-set, but the proto field was a non-optional `bool` — so any partial update (e.g. `given_name` only) sent the default `false` and turned MFA off. Fix: `optional bool is_multi_factor_auth_enabled` + map the `*bool` through; proto regenerated. Regression test added (`TestUpdateProfileGRPCDoesNotDisableMFA`). GraphQL was never affected.
2. **[High] AuthenticatorProvider missing from the service layer in the test harness** → any TOTP path through the service nil-panicked, and there was zero TOTP-through-service coverage. Fix: wired it in `test_helper.go`; added `TestVerifyOTPTOTPThroughService` (passcode + recovery code + invalid). Production wiring (`cmd/root.go`, `cmd/mcp.go`) was already correct.

### Nits (resolved)
- Restored sanitized error strings lost in the port: `verify_otp` passcode/recovery-code validation faults and `magic_link_login` update-user failure no longer surface the raw internal error; `resend_otp` keeps its response body.

### Verified safe (no action)
- Login timing-equalization (dummy bcrypt) + generic error collapse → no account enumeration; forgot_password/magic_link/resend_* enumeration-safe responses preserved; OTP HMAC compare + single-use + expiry intact; cookie flags (HttpOnly/Secure/SameSite/domain pair) preserved; the gateway `WithOutgoingHeaderMatcher` forwards only `Set-Cookie` and stops `Grpc-Metadata-*` leakage; typed-error mapping doesn't downgrade auth failures; credential-returning RPCs remain MCP-unexposed.
- One Info note (no action): `EmitUnpopulated` emits empty `authenticator_secret`/`recovery_codes` on REST AuthResponse — confirmed never real secret material, contract difference only vs GraphQL.

**Verdict:** security-engineer APPROVE WITH NITS; principal-engineer REQUEST CHANGES → all must-fix items now resolved and verified. Full `integration_tests` + `grpcsrv` + `gateway` + `service` suites green on SQLite; build/vet/gofmt/buf-lint clean.

PR #668, opened 2026-07-10
fix(oauth): cache OIDC discovery and fix Twitter PKCE

lakhansamani:
## Summary

Two fixes in the social-login flow (`internal/http_handlers/oauth_login.go`, `oauth_callback.go`).

### 1. Cache OIDC discovery documents (perf)

`oidc.NewProvider(ctx, issuer)` was called on **every** login for Google, Apple, Microsoft, and Twitch. Each call does a network round-trip to fetch the issuer's `/.well-known/openid-configuration` before any JWKS/verification work — adding latency to every login and load on the IdP for a document that is effectively static.

New `getOIDCProvider` (in `oidc_provider_cache.go`) memoizes `*oidc.Provider` by issuer URL in a package-level `sync.Map`, lazily populated on first use. No TTL — discovery docs essentially never change (marked with a `ponytail:` ceiling comment noting where to add refresh if an IdP ever rotates its discovery doc). Applied at all four call sites.

### 2. Fix broken Twitter PKCE, remove dead Roblox PKCE (bug)

`processTwitterUserInfo` and `processRobloxUserInfo` passed a `code_verifier` into the token exchange, but `AuthCodeURL` never sent a corresponding `code_challenge`. Worse, the value passed as the verifier was `sessionState` — the **provider name string** (`"twitter"` / `"roblox"`), not a real verifier. So the exchange presented a bogus verifier the provider was never given a challenge for.

Ground truth from provider docs:
- **Twitter/X** — [requires PKCE](https://developer.twitter.com/en/docs/authentication/oauth-2-0/authorization-code) for the OAuth 2.0 user-context auth code flow (`code_challenge` is mandatory). This is why the verifier code existed. → **Wired PKCE end-to-end**: generate a verifier at login (`oauth2.GenerateVerifier`), send the S256 challenge (`oauth2.S256ChallengeOption`), store the verifier keyed by state in the existing state store, retrieve it at the callback (`GetAndRemoveState`), and replay it in the exchange (`oauth2.VerifierOption`).
- **Roblox** — [requires PKCE only for public clients](https://create.roblox.com/docs/cloud/auth/oauth2-overview); optional for confidential clients. Authorizer always uses a client secret for Roblox (confidential), so PKCE is optional. → **Removed the dead `code_verifier` param.**

The other 8 providers are untouched (no PKCE change needed).

## Tests

- `oidc_provider_cache_test.go` — cache returns the same `*oidc.Provider` for the same issuer (fetching the discovery doc only once) and a different instance for a different issuer, using local `httptest` discovery servers.
- `oauth_pkce_test.go` — the S256 `code_challenge` sent at login matches `sha256(verifier)` of the replayed verifier, and the verifier state key is namespaced so it can't collide with the provider-state entry.

**Coverage note:** there is no full OAuth-callback integration test for Twitter or Roblox — they require live provider mocking, matching the existing pattern for the other 8 social providers (none have callback integration tests either). The PKCE fix is covered by the focused unit tests above.

## Verification

- `internal/http_handlers` package tests pass; `go vet` clean.
- OAuth integration tests (`oauth_account_linking`, `oauth_standards_compliance`) pass on SQLite.

### DIFF
--- a/internal/http_handlers/oauth_callback.go
+++ b/internal/http_handlers/oauth_callback.go
@@ -111,13 +111,20 @@ func (h *httpProvider) OAuthCallbackHandler() gin.HandlerFunc {
 		case constants.AuthRecipeMethodDiscord:
 			user, err = h.processDiscordUserInfo(ctx, oauthCode)
 		case constants.AuthRecipeMethodTwitter:
-			user, err = h.processTwitterUserInfo(ctx, oauthCode, sessionState)
+			// Twitter/X uses PKCE: retrieve the verifier stored at login keyed by state.
+			verifier, verr := h.MemoryStoreProvider.GetAndRemoveState(pkceVerifierKeyPrefix + state)
+			if verr != nil || verifier == "" {
+				log.Debug().Err(verr).Msg("Missing PKCE verifier for Twitter callback")
+				ctx.JSON(400, gin.H{"error": "invalid oauth state"})
+				return
+			}
+			user, err = h.processTwitterUserInfo(ctx, oauthCode, verifier)
 		case constants.AuthRecipeMethodMicrosoft:
 			user, err = h.processMicrosoftUserInfo(ctx, oauthCode)
 		case constants.AuthRecipeMethodTwitch:
 			user, err = h.processTwitchUserInfo(ctx, oauthCode)
 		case constants.AuthRecipeMethodRoblox:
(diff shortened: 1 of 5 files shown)


lakhansamani:
Fixed the nil-pointer-deref ordering finding: `oidcProvider.Verifier(...)` was called before the `err` check for Google (checked too late) and Microsoft (never checked). Apple and Twitch were already correctly ordered. Full test suite green.

PR #670, opened 2026-07-10
security(mfa): per-user TOTP lockout + hash recovery codes at rest

lakhansamani:
## Summary

Closes two confirmed gaps in the TOTP/MFA path (secrets are already AES-256-GCM encrypted at rest; these harden the remaining two).

### 1. Per-user verify_otp lockout (brute-force)
The only guard on failed OTP verification was the global per-IP rate limiter, so a single account could be brute-forced by spreading guesses across many IPs (or slowly under the limiter's radar). Add a per-user failed-attempt counter in the existing memory store (`SetCache`/`GetCache`/`DeleteCacheByPrefix`) — transient KV state, **no DB schema column**, so none of the 13+ storage providers change.

- 5 failed attempts within a 15-minute window locks TOTP/recovery verification for that user for 15 minutes.
- While locked, even a correct code is refused with a **distinct** error (`too many failed attempts...`, mapped to gRPC `ResourceExhausted` / HTTP 429) so a legitimate user understands why they're blocked — not the generic `invalid otp`.
- A successful verification resets the counter.
- A storage fault during recovery-code validation is **not** counted as a user failure (avoids self-lockout during an outage).

### 2. Recovery codes hashed at rest
Recovery codes were generated as random UUIDs but stored as **plaintext** in the `Authenticator.RecoveryCodes` JSON map. Now only their **SHA-256** hashes are persisted (these are already high-entropy — a slow hash like bcrypt/argon2 would be wasted cost).

- `Generate()` still returns the plaintext codes to the caller once for enrollment display (unchanged UX); only hashes are written.
- `ValidateRecoveryCode()` hashes the input and matches against stored hashes, marking the entry consumed exactly as before.
- **Lazy migration:** already-enrolled users with plaintext codes still validate once via a plaintext fallback (a stored key that isn't a 64-char hex hash is compared directly), mirroring the existing TOTP-secret migration in `internal/crypto/totp.go`. New enrollments always write hashes.

## Out of scope (deliberately)
Trusted-device / "remember this device" and per-client/per-org MFA policy — these need separate product design.

## Tests
- `internal/crypto/totp_test.go` — hash helpers (determinism, format detection).
- `internal/integration_tests/totp_recovery_codes_at_rest_test.go` — plaintext never persisted; hashed codes validate once; legacy plaintext codes validate once via fallback.
- `internal/integration_tests/verify_otp_totp_lockout_test.go` — 5 failures lock; a correct code is then refused with the lockout error; a success resets the counter.
- `make test-sqlite` green (30 packages, 0 failures); `make fmt-go` clean.

### DIFF
--- a/internal/authenticators/totp/totp.go
+++ b/internal/authenticators/totp/totp.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"image/png"
 	"time"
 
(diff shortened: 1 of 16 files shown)


lakhansamani:
Addressed both findings from the independent security review:

1. **Lockout counter race (HIGH)** — replaced the check-then-increment GetCache/SetCache pattern with a new atomic `IncrementCache` primitive (added to `memory_store.Provider`, implemented for all three backends). Now increment-then-check: concurrent callers get strictly increasing counts, so at most 5 requests can ever reach validation regardless of concurrency. Added a 20-concurrent-request regression test run under `-race`.
2. **Unauthenticated lockout DoS (MEDIUM)** — the MFA session is now validated before any TOTP/OTP/lockout-counter work runs, instead of after. An attacker who only knows a victim's email can no longer touch (and exhaust) their lockout counter.

Full test suite green (including the new concurrency test), two existing tests updated where they relied on the old ordering.

PR #686, opened 2026-07-15
feat(mfa): behavior redesign — withheld-token first-time setup, WebAuthn/Email/SMS as MFA factors, lockout + admin recovery

lakhansamani:
## Summary

Redesigns Authorizer's MFA behavior on top of the just-merged EnforceMFA/passkey-gate work (#685):

- **Token withholding**: a first-time optional-MFA offer no longer issues a token immediately then offers setup — the token is withheld until the user completes a method or explicitly skips, matching how enforced MFA already worked. Applies uniformly to password login, passkey login, signup, and OAuth/social login.
- **WebAuthn/passkey as a disableable MFA factor** (`--disable-webauthn-mfa`), folded into the `EnableMFA` derivation alongside TOTP/Email-OTP/SMS-OTP.
- **Email/SMS OTP as real, individually-enrollable MFA methods** — new `email_otp_mfa_setup`/`sms_otp_mfa_setup` mutations, with `verify_otp` now marking enrollment complete. Previously-existing login-time OTP branches (which fired for any optional-MFA user with a phone/email, no enrollment check at all) now require a verified enrollment first.
- **`effectiveMFAEnabled`**: `IsMultiFactorAuthEnabled` is no longer written as an inferred/persisted default at signup or login — it's either an explicit user/admin choice or computed live from server config, every call.
- **`skip_mfa_setup`** reworked from bearer-token to MFA-session-cookie authentication (no bearer token exists yet under the withheld-token model).
- **User-initiated lockout** (`lock_mfa`, refused when a verified Email/SMS OTP fallback exists) and **admin recovery** (`reset_mfa` on the existing `_update_user` mutation — clears all MFA state and deletes every enrolled factor across all 7 storage backends).
- **OAuth/social login gated identically to password login** (Google, GitHub, etc.) — verified against Auth0's actual documented behavior rather than assumed; enterprise SSO (SAML) and the routine session-refresh endpoint are untouched.

Design doc: `docs/superpowers/specs/2026-07-14-mfa-behavior-redesign-design.md` (local only, gitignored in this repo).
Plan: `docs/superpowers/plans/2026-07-14-mfa-behavior-redesign-backend.md`.

## Process

Built via subagent-driven development: 8 plan tasks, each implemented and independently reviewed (task-scoped spec + quality gate); a final whole-branch review across all 9 commits (opus) found no Critical issues but 2 Important findings, both fixed and re-reviewed:

- **WebAuthn-only enforced MFA silently bypassed password `Login()`/`SignUp()`** when TOTP was disabled — closed by broadening the gate to run whenever any MFA method is available, not just when TOTP specifically is.
- **Email/SMS OTP were offered on the first-time setup screen but structurally uncompletable there** (the enrollment mutations required a bearer token that doesn't exist yet in the withheld state) — closed by adding a cookie-authenticated fallback, proven end-to-end (withheld offer → cookie-authenticated setup → verify → token issued).

5 additional Minor findings (OAuth method-hint parity, defensive event-ordering, a weakened test assertion, missing SMS coverage, a small refactor) were also fixed and re-reviewed.

## Known follow-up (not blocking)

Verified-MFA-factor detection (`authenticatorVerified`) doesn't yet fold in Email/SMS-OTP for `webauthn.go` (any login channel) or for `login.go` in the cross-channel case (e.g. email-primary login with SMS-OTP enrolled as the MFA method) — an enrolled user in these cases is routed to re-enroll instead of verify. Confirmed non-security-bypassing (the token is still withheld either way); worth a follow-up ticket to fold Email/SMS-OTP into every `authenticatorVerified` computation site, matching what the OAuth gate now does.

## Test plan

- [x] `go build ./...`, `go vet ./...` clean
- [x] `internal/service`, `internal/integration_tests`, `internal/http_handlers`, `internal/config`, `internal/graph`, `internal/cookie` — all passing
- [x] Every task + every fix independently reviewed against its own diff, with security-critical tasks (token withholding, lockout, OAuth gate, the two Important-finding fixes) reviewed on a stronger model with independent re-tracing rather than trusting the implementer's own account
- [ ] Manual/browser verification of the real OAuth provider round-trip (no HTTP-level OAuth test double exists in this codebase; the gate's decision logic is covered by `TestEvaluateMFAGateForOAuth`, but the actual `/oauth_callback/:provider` redirect behavior with a live Google/GitHub login has not been driven end-to-end)
- [ ] Frontend (authorizer-js, authorizer-react) not yet updated — this PR is backend-only

## Update: proto/gRPC/REST parity

The redesign above shipped GraphQL-only. Added as follow-up commits on this same branch:
- Extended `AuthResponse`, `Meta`, `User` proto messages and admin's `UpdateUserRequest` with the fields the GraphQL layer already had (`should_offer_*_mfa_*`, `is_mfa_enforced`, `has_skipped_mfa_setup_at`, `mfa_locked_at`, `reset_mfa`).
- Added 4 new RPCs (`SkipMfaSetup`, `LockMfa`, `EmailOtpMfaSetup`, `SmsOtpMfaSetup`) with REST routes auto-generated via the existing gRPC-gateway, mirroring `VerifyOtp`'s exact pattern — pure transport wiring, zero `internal/service` changes.
- Also fixed `Meta.is_webauthn_enabled`, which predated `--disable-webauthn-mfa` and was hardcoded `true`.
- End-to-end gRPC tests for all 4 new RPCs, both auth modes where applicable (bearer token and MFA-session-cookie).
- Known gap noticed, not fixed (pre-existing, unrelated to this PR): `AuthResponse.should_offer_mfa_setup` (the deprecated, unqualified field) has no proto counterpart either.


lakhansamani:
Pushed a follow-up commit closing two gaps left open by this PR's earlier work:

1. **Passkey enrollment was never actually offered during the MFA setup screen** despite `should_offer_webauthn_mfa_setup` already signaling it was available — `webauthn_registration_options`/`_verify` required a bearer token that doesn't exist during a token-withheld offer. Added a fallback to the MFA session cookie (same pattern as `VerifyOTP`/`SkipMFASetup`), gated strictly to the offer/enroll gate states so a password-only session can never mint a new passkey to bypass challenging an *existing* second factor. `webauthn_registration_verify`'s GraphQL return type changes `Response` → `AuthResponse` to carry the token issued on completion (message-only, unchanged, for the ordinary settings-page caller).
2. **OAuth/magic-link returns couldn't distinguish "first-time offer" from "challenge an existing factor"** — `Root.tsx` always rendered the setup screen. Added `mfa_gate=offer|verify` to the redirect and branched accordingly (paired with authorizerdev/authorizer-react#63 / authorizerdev/authorizer-js#47).

Given the new session-cookie auth path on a credential-registration endpoint, flagging for `security-engineer` review specifically on `resolveWebauthnSetupCaller` (internal/service/webauthn.go) before merge.

Full Go test suite green, 5 new tests added covering the session path (happy path + the blockVerify bypass guard + missing-session + challenge-session + settings-page-unaffected regression). Passkey-setup-screen fix and email-OTP-via-mailpit flow verified live against a local build; `mfa_gate` verify/offer routing verified via both branches in-browser.

